PR #1, opened 2019-08-21
String 2 - Problem 1 & 2

Hariae:


jaspindersingh83 (commented):
Good job. We will optimize this one using KMP algo

PR #2, opened 2019-08-21
Completed Strings - 2 - Please Review

bharatreddypothireddy:


super30admin (commented):
Think about the time complexity again for this solution. Aren't we iterating m times (m length of needle) every time at every index in haystack. This is a brute force solution. Can you think of an optimal one ?

super30admin (commented):
Great job with problem 2

jaspindersingh83 (commented):
Problem 1 your time complexity is O(mn). We can get O(m+n) with KMP algorithm

PR #5, opened 2019-08-22
Problem 1 done, please review

Ameya145:


jaspindersingh83 (commented):
The time complexity is O(mn) actually. m --len of needle. n -- len of haystack. We can optimize it to O(m+n) using KMP

PR #6, opened 2019-08-22
Please review strings 2

AkhilKrishnamurthy:


jaspindersingh83 (commented):
This works but is a brute force solution. We can use KMP algorithm to optimize it

PR #14, opened 2019-09-18
Problem-2

Rohanp1196:
Problem-2


sehaj-grover (commented):
rohan try solving the problem on your own first and also add the details in comments

PR #17, opened 2019-10-09
Ststr implement using KMP. LC verified

manishjangid:


kaurramanpreet (commented):
great work manish 
the way you explain your code is amazing!!

PR #21, opened 2019-11-26
Problems 1 and 2

keerthitejakonuri:
Problems 1 and 2(TLE for 2nd problem for few testcases)


sehaj-grover (commented):
Very good Keerthi
Do try the solutions after the session as well

PR #22, opened 2019-11-26
Completed strStr

Ajayvikram10:


sehaj-grover (commented):
AJay please add details to the code 
And make sure to try the problems after the session

PR #47, opened 2020-03-03
pls review.

Nsharma335:


problem-1
```diff
@@ -0,0 +1,36 @@
...
+              }
+              if(j==m)
+                  return i;
+             
+          }
+           i++;
+    }
+        return -1;
+}
+}
```
rajatsharma-pau, line 36: The optimal solution for this problem is below
class Solution {
        int m = haystack.length();
        int n = needle.length();
        if(n == 0) return 0;
        int [] lps = lps(needle, n);
        int i = 0; int j = 0;
        while(i < m){
            if(haystack.charAt(i) == needle.charAt(j)){
                i++; j++;
            }
            if(j == n) {
                return i - j;
            } else if(i < m && j > 0 && haystack.charAt(i) != needle.charAt(j)){
                j = lps[j-1];
            } else if (i < m && j == 0 && haystack.charAt(i) != needle.charAt(j)){
                i++;
            }
        }
        return -1;
        
    }
    private int [] lps (String pattern, int n){
        int [] lps = new int [n];
        int i = 1;
        int j = 0;
        lps[0] = 0;
        while(i < n){
            // Case 1
            if(pattern.charAt(j) == pattern.charAt(i)){
                j++; lps[i] = j; i++;
            } else if (j > 0 && pattern.charAt(j) != pattern.charAt(i)){
                j = lps[j-1];
            } else if (j == 0 && pattern.charAt(j) != pattern.charAt(i)){
                lps[i] = 0;
                i++;
            }
        }
        return lps;
    }
}

rajatsharma-pau (commented):
Neha Good to see that Anagrams problems  you have solved both ways one is giving TLE and others solution is optimized way.

The second solution for probem strStr() is  to implement with brute force is fine . I have also added the optimal solution  in the same file

rajatsharma-pau:
Neha Good to see that Anagrams problems  you have solved both ways one is giving TLE and others solution is optimized way.

The second solution for problem strStr() is  to implement with brute force is fine . I have also added the optimal solution  in the same file

PR #55, opened 2020-03-25
initial version strstr() - Please review

priyanka1012-singhal:


Problem1.java
```diff
@@ -0,0 +1,15 @@
+// Problem1  Implement strStr() (https://leetcode.com/problems/implement-strstr/)
+
+//Time comlexity O(hlen (hlen - nlen)) 
+class Solution {
```
super30admin, line 4: its m*n brute force. For each index in haystack you are checking if the substring of needle is there or not. Try KMP

super30admin (commented):
its m*n brute force. For each index in haystack you are checking if the substring of needle is there or not. Try KMP

super30admin (commented):
Just saw . KMP is correct

super30admin:
Just saw . KMP is correct

PR #58, opened 2020-03-26
Partially Done.

saumil0257:


solution.md
```diff
@@ -0,0 +1,71 @@
...
+              while end<len(s):
+                  ch=s[end]
+                  freqWindow[ord(ch)-ord('a')]+=1
+                  if end-start+1 > len(p):
+                      freqWindow[ord(s[start])-ord('a')]-=1
+                      start+=1
+              if freqWindow==freqP:
+                  #CHecking anagram
+                  result.append(start)
+              end+=1
```
super30admin, line 40: There is an indentation issue here.
`if freqWindow==freqP:
                  #CHecking anagram
                  result.append(start)
              end+=1`
is missing one index making it out of the loop

super30admin (commented):
Fixed code for problem 1
```class Solution:
              #Edge Case
              if len(s)<len(p):
                  return
              freqP=[0]*26
              freqWindow=[0]*26
              #Poppulate freqP
              for i in range(len(p)):
                  ch=p[i]
                  #print(freqP[ord(ch)-ord('a')])
                  freqP[ord(ch)-ord('a')]+=1
              start=end=0
              result=[]
              while end<len(s):
                  ch=s[end]
                  freqWindow[ord(ch)-ord('a')]+=1
                  if end-start+1 > len(p):
                      freqWindow[ord(s[start])-ord('a')]-=1
                      start+=1
                  if freqWindow==freqP:
                      #CHecking anagram
                      result.append(start)
                  end+=1
              return result```

2> Try KMP for strStr. Please don't miss this one

super30admin:
Please review inline comments

PR #68, opened 2020-05-06
Completed Haystack Needle question: Please review

KBangera:


Haystack_Needle.java
```diff
@@ -0,0 +1,25 @@
...
+        int m = haystack.length();
+        int n = needle.length();
+        
+        if(m < n)
+            return -1;
+        else if(m == n)
+            return haystack.equals(needle)?0:-1;
+        else{
+            if(haystack.contains(needle)){
+                return haystack.indexOf(needle);
```
devashishnyati, line 20: indexOf is O(n) operation.

Haystack_Needle.java
```diff
@@ -0,0 +1,25 @@
+// Time Complexity :O(n)
```
devashishnyati, line 1: Worst case this should be O(n^2), right?

devashishnyati (commented):
Good try Karthik. 
Try KMP solution also.
Also, submit the other problem.

devashishnyati:
Good try Karthik. 
Try KMP solution also.
Also, submit the other problem.

PR #69, opened 2020-05-06
adds the code for strings2

shahakshay11:


devashishnyati (commented):
Good job Akshay!
The strstr solution looks different. Try KMP algorithm also.

devashishnyati:
Good job Akshay!
The strstr solution looks different. Try KMP algorithm also.

PR #100, opened 2020-07-22
Completed Strings-2. Kindly review and share feedback.

Sumesh-H:


AroopN:
Good work, Sumesh. Another interesting way to solve implement strstr() in O(N) is by using rolling hash. Look that up.

AroopN:
Good work, Sumesh. Another interesting way to solve implement strstr() in O(N) is by using rolling hash. Look that up.

Sumesh-H:
> Good work, Sumesh. Another interesting way to solve implement strstr() in O(N) is by using rolling hash. Look that up.

I will try but I think it will give TLE. The best approach here is KMP correct?

AroopN:
> > Good work, Sumesh. Another interesting way to solve implement strstr() in O(N) is by using rolling hash. Look that up.
> 
> I will try but I think it will give TLE. The best approach here is KMP correct?

Yes, KMP will be the more optimal solution though both KMP as well as rolling hash have O(n) TC.
Though rolling hash wont give TLE, Rolling hash shouldn't be generally used as a primary solution for this problem but can be used when asked for an alternative optimized solution.

Sumesh-H:
> > > Good work, Sumesh. Another interesting way to solve implement strstr() in O(N) is by using rolling hash. Look that up.
> > 
> > 
> > I will try but I think it will give TLE. The best approach here is KMP correct?
> 
> Yes, KMP will be the more optimal solution though both KMP as well as rolling hash have O(n) TC.
> Though rolling hash wont give TLE, Rolling hash shouldn't be generally used as a primary solution for this problem but can be used when asked for an alternative optimized solution.

Alright, I will try this approach and share the solution.

PR #203, opened 2021-02-10
Both problems done

nilesh-525:
strstr brute force is written. I will update PR when KMP is done.


khushal1198 (commented):
Good work! Please update strstr soon as you mentioned.

khushal1198:
Good work! Please update strstr soon as you mentioned.

PR #207, opened 2021-02-11
Completed Strings -2. Please review

gayathrig269:


kannan-mrugapphan (commented):
Good Job. Please recheck on the needle and haystack problem. It runs at n^2 time. Try to optimize it using KMP algorithm.

kannan-mrugapphan:
Good Job. Please recheck on the needle and haystack problem. It runs at n^2 time. Try to optimize it using KMP algorithm.

PR #293, opened 2021-08-01
added strings 2

pandas5308:


maithrreye:
Seems good ! Its hard to decipher a code without comments . Please add complexities too !

maithrreye:
Seems good ! Its hard to decipher a code without comments . Please add complexities too !

PR #412, opened 2022-06-17
done

shilpa271:


niharikagupta2021 (commented):
Good Job! Please complete the other problem as well and include sc and tc for both problems

niharikagupta2021:
Good Job! Please complete the other problem as well and include sc and tc for both problems

PR #565, opened 2023-06-09
Strings-2 Solutions

thanujkumarj:


problem2.cpp
```diff
@@ -0,0 +1,41 @@
+// Time Complexity :O(m+n)
```
ishika2741, line 1: Kindly tell what is m and n

ishika2741 (commented):
Well done! Keep it up.

ishika2741:
Well done! Keep it up.

problem2.cpp
thanujkumarj: m = p.size() , n = s.size()

PR #569, opened 2023-07-12
Completed Strinngs 2

sanketauti8:


438 Find All Anagrams in a String.txt
```diff
@@ -0,0 +1,45 @@
+//438. Find All Anagrams in a String
+
```
ishika2741, line 2: Add TC and SC for this problem.

28 Find the Index of the First Occurrence in a String.txt
```diff
@@ -0,0 +1,28 @@
+//28. Find the Index of the First Occurrence in a String
+
+//TC=)(n*m)
```
ishika2741, line 3: Try to submit rolling hash solution as well.

ishika2741 (commented):
Good work! Keep it up.

ishika2741:
Good work! Keep it up.

PR #617, opened 2023-09-22
Complete String 2 - Please Review

saivinherkar:


Strstr.java
```diff
@@ -0,0 +1,21 @@
+// TC = O(mn)
```
ishika2741, line 1: submit this solution with KMP and robin karo algo as well.

ishika2741 (commented):
Good work Sai!

ishika2741:
Good work Sai!

PR #679, opened 2024-03-17
problems solved

kpranavreddy01:


Nandha2Tech (commented):
Great Work! For anagrams problem TC would be O(m+n) and SC would be O(n). For finding the first occurrence problem TC would be O(m*n) and SC would be O(1). Try to do Rabin Karp's algo or KMP algo for the second problem.

Nandha2Tech:
Great Work! For anagrams problem TC would be O(m+n) and SC would be O(n). For finding the first occurrence problem TC would be O(m*n) and SC would be O(1). Try to do Rabin Karp's algo or KMP algo for the second problem.

PR #711, opened 2024-08-07
completed strings-2

dhikki-study:
##28. Find the Index of the First Occurrence in a String##################################################
// Time Complexity : m+n -> length or substrate + length of string
// Space Complexity : O(1) -> no extra space just calculation for rolling hash
// Did this code successfully run on Leetcode : Yes
// Any problem you faced while coding this : no problem

// Your code here along with comments explaining your approach in three sentences only
Here we used rolling has to identify the substring value and than we used sliding window to find rolling hash value for each substring with the length equal to matching string, while moving I made sure to remove previous element by taking mod with 10^len(matchstring-1) 

##438. Find All Anagrams in a String################################################################
// Time Complexity : m+n
// Space Complexity : O(1) -> used to maintain the dictionary
// Did this code successfully run on Leetcode : Yes
// Any problem you faced while coding this : no problem

// Your code here along with comments explaining your approach in three sentences only
We create a dictionary match string where count of each element is stored, now we apply sliding window on main string and when a match is found we reduce the count of element form dict and also make sure our match index is increased when any element count drops to zero. When the match count reaches the length of dict it means we have found match. When performing the sliding window we also make sure to add count to dict when any element is removed and also reduce match count when any element count change for 0 to 1

### DIFF
--- a/28. Find the Index of the First Occurrence in a String.py
+++ b/28. Find the Index of the First Occurrence in a String.py
@@ -1,0 +1,35 @@
+##28. Find the Index of the First Occurrence in a String########################################################################################
+// Time Complexity : m+n -> length or substrate + length of string
+// Space Complexity : O(1) -> no extra space just calculation for rolling hash
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : no problem
+
+// Your code here along with comments explaining your approach in three sentences only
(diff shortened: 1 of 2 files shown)


Nandha2Tech (commented):
Great Work!

Nandha2Tech:
Great Work!

PR #716, opened 2024-08-09
Done Strings-2

ssd9881:


Nandha2Tech (commented):
Great Work! TC for strstr would be O(m*n).

Nandha2Tech:
Great Work! TC for strstr would be O(m*n).

PR #729, opened 2024-09-17
done

akash-shindhe86:


Erlybird:
Problem 1 is not working on leetcode because it's going out of bounds for long type in a test case. Check the problem and resubmit. But logic is correct.

Erlybird:
Problem 1 is not working on leetcode because it's going out of bounds for long type in a test case. Check the problem and resubmit. But logic is correct.

PR #790, opened 2025-03-01
completed Strings-2

harshadatupe:


find-the-index-of-the-first-occurrence-in-a-string.py
```diff
@@ -0,0 +1,8 @@
+# tc O(m * n), sc O(1).
```
ishika2741, line 1: Please submit the optimal solution as well

ishika2741 (commented):
Good work!

ishika2741:
Good work!

PR #870, opened 2025-07-29
solved using running hash(string pattern)

IshaaS:


super30admin:
- Strengths:
  - The student demonstrates a good understanding of both naive and optimized approaches for string matching problems.
  - The code is well-commented and easy to follow.
  - The solutions are correct and handle edge cases appropriately (e.g., checking if `needle` is longer than `haystack`).

- Areas for Improvement:
  - For the hashing solution in `strStr`, consider using a simpler rolling hash with primitive types to avoid the overhead of `BigInteger`.
  - While the naive `strStr` implementation is correct, it might be worth mentioning or implementing more advanced algorithms like KMP for educational purposes.
  - The `Sample.java` file is empty and should either be removed or completed to avoid confusion.

PR #871, opened 2025-08-19
Done Strings-2

NUPUR0307:


super30admin:
Strengths:
- Good problem-solving approach and clear explanations in comments.
- Correct handling of edge cases in Problem2.
- Efficient use of sliding window technique in Problem2.

Areas for Improvement:
- In Problem1, the edge case handling for empty strings needs correction. When needle is empty, return 0, but when haystack is empty and needle is not, return -1.
- For Problem1, consider using character-by-character comparison instead of substring for slightly better performance, though the current approach is acceptable.
- The README.md file should be separate from the code files. It's better to have the code in separate .java files and the explanation in the README.

PR #872, opened 2025-08-20
Completed Strings-2

srilekhad:


super30admin:
- The student's solution is correct and efficient for the given problem. The implementation is clean and easy to understand.
- The comments provided are helpful and demonstrate a good understanding of the problem and solution.
- One minor improvement could be to reset `j` to 0 immediately after the inner loop breaks, rather than waiting until the end of the outer loop iteration. This is a small optimization but doesn't affect the overall complexity.
- The student could also consider adding a brief comment explaining why the loop runs until `i <= m - n`, to clarify the boundary condition for future readers.

PR #875, opened 2025-09-02
solved 2 questions

nsh2207:


super30admin:
Strengths:
- The implementation of the KMP algorithm in Question_1.cpp is excellent and demonstrates a deep understanding of efficient string matching.
- Both solutions are well-commented and easy to follow, showing good coding practices.
- The use of sliding window in Question_2.cpp is appropriate for the problem.

Areas for Improvement:
- In Question_2.cpp, the frequency comparison can be optimized. Instead of comparing the entire frequency maps, you could maintain a count of matched characters, which would reduce the time complexity to O(n).
- The comments in Question_2.cpp mention O(n*m) time complexity, but it's important to note that this can be improved to O(n) with a better approach.
- Consider adding edge case checks (e.g., empty strings) to make the solutions more robust.

PR #876, opened 2025-09-11
Strings-2 completed

Swathi-L:


super30admin:
- Strengths:
  - Both solutions are correct and efficient.
  - The use of comments and helper methods improves code readability.
  - The sliding window technique in `FindAnagrams.java` is well-implemented.
- Areas for Improvement:
  - For `StrStr.java`, consider mentioning the possibility of using more advanced algorithms like KMP for better theoretical performance, though the current solution is practical for many cases.
  - In `FindAnagrams.java`, ensure that the hashmap size check (`match == pFreqMap.size()`) is clearly understood; it might be helpful to add a comment explaining why this condition works.

PR #881, opened 2025-09-24
added solution for Strings-2

satvikapadakanti:


super30admin:
- Strengths:
  - The solution is correct and handles edge cases well (e.g., empty needle, needle longer than haystack).
  - The code is concise and leverages built-in methods effectively.
  - The logic is straightforward and easy to follow.

- Areas for Improvement:
  - Consider adding a comment to explain the loop condition `haystack.length() - needle.length() + 1` to make it clearer for readers.
  - While the current approach is correct, manually comparing characters (like in the reference solution) could be slightly more efficient in some cases, though the difference is minimal.
  - Ensure consistent indentation and formatting for better readability.

PR #883, opened 2025-10-08
Strings-2 completed

sathvikareddy18:


super30admin:
Strengths:
- Good use of sliding window technique in `findAnagrams`.
- Attempt at an optimized solution (Rabin-Karp) in `strStr` shows understanding of advanced algorithms.
- Generally clean code structure in both solutions.

Areas for Improvement:
- The `strStr` implementation has a critical error in the hash calculation. When removing the outgoing character, you need to subtract its contribution, not add it. The line should be `curr_hash = curr_hash - (PosFac * (ord(c_out) - ord('a') + 1))`.
- Consider adding comments to explain the hashing approach in `strStr` for better readability.
- The variable `PosFac` should be named more descriptively (perhaps `power` or `highestPower`).
- More thorough testing would have caught the hashing error in `strStr`.

PR #884, opened 2025-10-10
Add solutions for anagram and substring search problems

ravithakur2k:
Added FindAllAnagrams.py implementing an efficient sliding window approach to find all anagrams of a pattern in a string. Added ImplementStrStr.py with two methods for substring search: a two-pointer approach and a Rabin-Karp based hashing solution.


super30admin:
Strengths:
- Correct implementation of both naive and Robin-Karp algorithms
- Good understanding of time and space complexity
- Includes an optimized solution (Robin-Karp) showing deeper knowledge

Areas for Improvement:
- The `findAnagrams` solution is misplaced and should be in a separate file
- Variable names could be more descriptive (e.g., `start_index` instead of `k`)
- The `else` clause in the `strStr` implementation is unnecessary since the loop will break naturally
- Comments could be more detailed, especially explaining the Robin-Karp implementation
- The `findAnagrams` solution's time complexity is incorrectly stated as O(n) - it should be O(m) where m is the length of s

PR #885, opened 2025-10-10
Strings-2-Done

candida17:


super30admin:
Strengths:
- Good understanding and implementation of advanced string matching algorithms (Rabin-Karp and sliding window with hash map).
- Clear and well-commented code that explains the approach.
- Appropriate variable naming and code structure.

Areas for Improvement:
- For Problem1.java, consider adding a character-by-character comparison when hashes match to handle potential hash collisions (though rare with this implementation).
- The comment in Problem1.java about not passing all test cases suggests there might be an edge case not handled - it would be good to identify and address that specific case.
- While the space complexity is technically O(1) for Problem2.java, it might be worth mentioning that it's O(26) specifically since we're dealing with lowercase English letters.

PR #888, opened 2025-10-16
Strings-2

mdashora:
Implement strStr method to find the index of the first occurrence of a substring.

### DIFF
--- a/Solution
+++ b/Solution
@@ -1,0 +1,33 @@
+// TC - O(mxn)
+// SC- O(1)
+
+class Solution {
+    public int strStr(String haystack, String needle) {
+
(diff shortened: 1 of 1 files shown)


super30admin:
- Strengths:
  - The solution is correct and handles edge cases well.
  - The code is clean, readable, and well-commented.
  - The time and space complexity are optimal and match the reference solution.
- Areas for Improvement:
  - Consider optimizing the outer loop condition to `i <= m - n` to avoid unnecessary iterations when the remaining haystack length is less than the needle length. This small change can improve efficiency in some cases.
  - While the current implementation is correct, adding a check for an empty needle (though the problem might not require it) could make the solution more robust.

PR #890, opened 2025-10-16
Completed Strings 2

Saideepthi123:


super30admin:
- Strengths:
  - The `findAnagrams` solution is well-implemented, efficient, and correctly uses a sliding window technique with a hash map to track character frequencies.
  - The code is well-commented and easy to follow, demonstrating good understanding of the problem and solution approach.

- Areas for Improvement:
  - The `strStr` solution is incorrect due to the flawed hashing approach. Hashing strings with a base of 26 can lead to collisions, especially for longer strings, resulting in false positives. Consider using a more robust hashing method or reverting to the brute-force approach.
  - The time complexity analysis for `strStr` is misleading because the solution doesn't work as intended. Ensure that the solution is correct before analyzing its complexity.

PR #893, opened 2025-10-27
Strings-2 done

neelp1025:


super30admin:
Strengths:
- The student has provided multiple solutions with different approaches, showing a good understanding of the problem.
- The code is well-commented and easy to follow.
- The use of rolling hash in HashingSolution is a clever optimization.
- The solution for Problem2 is particularly well-implemented with clear logic for handling incoming and outgoing characters.

Areas for Improvement:
- For HashingSolution, consider if BigInteger is necessary or if a simpler hash function (with modulo to prevent overflow) could be used for typical problem constraints.
- The HashingSolution could benefit from a brief explanation of why base 26 is chosen (since there are 26 letters in the alphabet).
- In Problem2, the initial check for `m < n` could be moved to the beginning of the method for better readability.

PR #897, opened 2025-11-20
Completed Strings 2

restacksyj:


super30admin:
Strengths:
- Good use of algorithms (rolling hash and sliding window) to solve the problems.
- Clear comments explaining the approach.
- Efficient solutions with optimal time and space complexity.

Areas for Improvement:
- In Problem1.cpp, consider adding a verification step or using double hashing to handle collisions.
- Avoid using `#include <bits/stdc++.h>` in favor of specific headers needed.
- For Problem1.cpp, the time complexity claim should be clarified to account for potential worst-case scenarios due to hash collisions.

PR #898, opened 2025-11-20
Strings-2-Complete

PrasiddhShah:


super30admin:
- Strengths:
  - The student demonstrates a good understanding of both problems and provides correct solutions.
  - The use of comments to explain the approach is excellent and aids in understanding the code.
  - The code is clean, well-organized, and follows best practices.

- Areas for Improvement:
  - For `strStr`, while the solution is correct, it might be worth mentioning that more advanced algorithms like KMP or Boyer-Moore exist for better average-case performance, though they are more complex.
  - In `findAnagrams`, the comment could be slightly more detailed to explain why the match count is compared to the map size, which might not be immediately obvious to all readers.

PR #899, opened 2025-11-21
Strings-2 solutions

kalyan-ng22:


super30admin:
- The student's solution is well-implemented and follows best practices. The comments and variable names enhance readability.
- The logic is straightforward and correctly handles the problem requirements.
- One minor improvement could be to add a check at the beginning to handle the case where the needle is an empty string, though this is often handled by the problem constraints.
- The student could consider mentioning the possibility of more advanced algorithms (like KMP or Boyer-Moore) in the comments for future reference, even if they are not implemented here.

PR #902, opened 2026-01-06
added the solution

anilkumarm0921-oss:
String2


super30admin:
- Problem1.java: Your solution is correct and efficient, but you can optimize it further by breaking early when characters don't match. This would reduce the number of unnecessary comparisons and improve performance in some cases. The count variable is not necessary; you can directly compare characters and break if they don't match.
- Problem2.java: Your solution is excellent. It correctly uses a sliding window approach with a hashmap to efficiently find anagrams. The code is clean and well-structured. Keep up the good work!

PR #903, opened 2026-02-11
Find Anagrams Solution

dbkuppagiri:


super30admin:
It seems there has been a confusion in the problem you are solving. The problem you were asked to solve is "strStr", which requires finding the first occurrence of a substring (needle) in a string (haystack). However, your solution is for "findAnagrams", which is a different problem that finds all starting indices of anagrams of a string (p) in another string (s).

For the strStr problem, you need to implement a function that takes two strings: haystack and needle, and returns the index of the first occurrence of needle in haystack, or -1 if not found. A straightforward approach is to use a nested loop to check each possible starting index in haystack for the substring needle. Alternatively, you can use more advanced algorithms like KMP for better worst-case performance, but the simple approach is acceptable for most cases.

Here is a simple solution for strStr in JavaScript:

```javascript
var strStr = function(haystack, needle) {
    const n = haystack.length;
    const m = needle.length;
    if (m === 0) return 0;
    for (let i = 0; i <= n - m; i++) {
        let j = 0;
        while (j < m && haystack[i + j] === needle[j]) {
            j++;
        }
        if (j === m) return i;
    }
    return -1;
};
```

Please ensure you are solving the correct problem in the future. Double-check the problem statement and examples before coding.

super30admin:
Your solution for the findAnagrams problem is well-implemented with a sliding window and frequency map. The logic is correct, and the time and space complexities are optimal for that problem. However, you have submitted this solution for the strStr problem, which requires a different approach. 

For the strStr problem, you need to find the first occurrence of a substring (needle) in a string (haystack). A common approach is to use a nested loop to compare each possible starting position in haystack with the needle string. Alternatively, you could implement more advanced algorithms like KMP for better worst-case performance, but the nested loop solution is acceptable for most cases.

Please ensure you are solving the correct problem. For strStr, you should write a function that takes two parameters: haystack and needle, and returns the index of the first occurrence of needle in haystack.

Here is a simple example of how you might implement strStr in JavaScript:
```
var strStr = function(haystack, needle) {
    if (needle.length === 0) return 0;
    for (let i = 0; i <= haystack.length - needle.length; i++) {
        if (haystack.substring(i, i + needle.length) === needle) {
            return i;
        }
    }
    return -1;
};
```
This solution checks each possible starting index in haystack by comparing substrings. Note that it has a time complexity of O(m*n) in the worst case, which is acceptable for the problem constraints.

PR #904, opened 2026-02-11
HW Problem 2 Find Anagrams

Amaan-29:


super30admin:
**Strengths**:
- The student understood the problem and attempted a common approach.
- The code includes comments that explain the intention.
- The student provided a main method for testing, which shows practical thinking.

**Areas for Improvement**:
1. **Syntax and Logic Error**: The extra semicolon after the inner if condition is a critical mistake. It should be removed. The correct code should be:
   ```java
   if (haystack.substring(i, needle.length() + i).equals(needle)) {
       return i;
   }
   ```
   Alternatively, the student can use a loop to compare characters without creating a substring to avoid potential overhead.

2. **Edge Cases**: The code does not handle the case when needle is empty. According to the problem, we should return 0 when needle is empty. The current code would throw an exception when needle.length() is 0 because needle.charAt(0) would be called. So, we need to check if needle is empty at the beginning.

3. **Efficiency Consideration**: Using substring and equals might be acceptable, but it's better to compare characters without creating a substring to save memory and time, especially for large strings. The reference solution does this by using a while loop to compare each character.

4. **Code Structure**: The solution should be in a class named Solution to match the typical LeetCode submission format. The main method is not required for online judges.

5. **Variable Names**: The variable names are appropriate, but the code could be more concise.

### Revised Solution Suggestion:
```java
public class Solution {
        if (needle.isEmpty()) return 0;
        int n = needle.length();
        int m = haystack.length();
        for (int i = 0; i <= m - n; i++) {
            if (haystack.charAt(i) == needle.charAt(0)) {
                int j = 0;
                while (j < n && haystack.charAt(i + j) == needle.charAt(j)) {
                    j++;
                }
                if (j == n) return i;
            }
        }
        return -1;
    }
}
```

###

PR #905, opened 2026-02-13
Completed Strings 2 problems

thondaha:
This PR consists of algorithms for 2 problems 
1. [Find All Anagrams in a String](https://leetcode.com/problems/find-all-anagrams-in-a-string/)
2.[Find the Index of the First Occurrence in a String](https://leetcode.com/problems/find-the-index-of-the-first-occurrence-in-a-string/)
3. Test class to test both the algorithms

### DIFF
--- a/src/AnagramsInAString.java
+++ b/src/AnagramsInAString.java
@@ -1,0 +1,51 @@
+/*
+Problem - Given two strings s and p, return an array of all the start indices of p's anagrams in s.
+Approach - So we count the characters of p in a map first.
+Then we move a sliding window on s and keep adjusting counts while checking matches.
+If all characters match in count, we store the starting index.
+Time Complexity - O(m+n)
+Space Complexity - O(1) // character map
+ */
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+
+public class AnagramsInAString {
+    public List<Integer> findAnagrams(String s, String p) {
+        List<Integer> result = new ArrayList<>();
+        int m = s.length();
+        int n = p.length();
+        if(n>m) return result;
+        HashMap<Character, Integer> map = new HashMap<>();
+        for(char c : p.toCharArray()){
+            map.put(c,map.getOrDefault(c,0)+1); // map the characters and their counts for a string P
+        }
+        int match = 0;
+        for(int i = 0; i < m; i++){
+            char in = s.charAt(i);
+            if(map.containsKey(in)){
+                int freq = map.get(in);
+                freq--;
+                map.put(in,freq);
+                if(freq == 0) {
+                    match++; // if a character of P is found in S increase the match count
+                }
+            }
+            if(i>=n){ // window size until length of string p
+                char out = s.charAt(i-n);
+                if(map.containsKey(out)) {
+                    int freq = map.get(out);
+                    freq++;
+                    map.put(out, freq);
+                    if (freq == 1) {
+                        match--;
+                    }
+                }
+            }
+            if(match == map.size()){
+                result.add(i-n+1); // starting index where the match is found
+            }
+        }
+        return result;
+    }
+}
--- a/src/FirstOccuranceOfAString.java
+++ b/src/FirstOccuranceOfAString.java
@@ -1,0 +1,45 @@
+/*
+Problem - Given two strings needle and haystack, return the index of the first occurrence of needle in haystack, or -1 if needle is not part of haystack.
+Approach - We compute a base-26 hash of the needle using BigInteger to prevent overflow.
+We roll a BigInteger hash window over the haystack and subtract the outgoing character's contribution when needed.
+If the rolling hash matches needle's hash, we return the starting index.
+Time Complexity - O(m+n)
+Space Complexity - O(1)
+ */
+import java.math.BigInteger;
+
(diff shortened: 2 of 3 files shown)


super30admin:
Strengths:
- You attempted to implement the Rabin-Karp algorithm, which shows knowledge of advanced string matching techniques.
- You included a substring check to handle hash collisions, which is a good practice.
- Your code is commented and structured.

Areas for Improvement:
- The Rabin-Karp implementation has a critical error in the hash calculation. The factor for the outgoing character should be `base^(n-1)`, not `base^n`. For example, if n=3, the hash for the first window is `c0 * base^2 + c1 * base^1 + c2 * base^0`. When moving to the next window, we subtract `c0 * base^2` and then multiply the remaining hash by base and add the new character. So you need to precompute `base^(n-1)`.
- Using BigInteger might be overkill and slow. Consider using modulo arithmetic with a large prime to avoid overflow and use long or int for faster operations.
- The space complexity can be reduced by not using BigInteger. With modulo arithmetic, you can use a long variable, which is constant space.
- The current solution does not handle the case when n=0 (needle is empty) correctly because the code returns 0 only if n==0 is checked. In your code, you have a check for n==0, but it is at the beginning. However, the problem constraints say that needle.length>=1, so it might not be necessary, but it's good to have.
- The variable names could be improved for clarity. For example, use `patternHash` instead of `pHash`, and `currentHash` instead of `cur`.

Suggested Correction for Rabin-Karp:
```java
    int m = haystack.length();
    int n = needle.length();
    if (n == 0) return 0;
    if (m < n) return -1;
    long base = 26;
    long mod = (long)1e9+7; // a large prime to avoid overflow
    long patternHash = 0;
    long currentHash = 0;
    long power = 1;
    for (int i = 0; i < n; i++) {
        patternHash = (patternHash * base + (needle.charAt(i) - 'a' + 1)) % mod;
        currentHash = (currentHash * base + (haystack.charAt(i) - 'a' + 1)) % mod;
        power = (power * base) % mod;
    }
    if (currentHash == patternHash && haystack.substring(0, n).equals(needle)) {
        return 0;
    }
    for (int i = n; i < m; i++) {
        currentHash = (currentHash * base + (haystack.charAt(i) - 'a' + 1)) % mod;
        currentHash = (currentHash - (haystack.charAt(i - n) - 'a' + 1) * power % mod + mod) % mod;
        if (currentHash == patternHash && haystack.substring(i - n + 1, i + 1).equals(needle)) {
            return i - n + 1;
        }
    }
    return -1;
}
```
Note: The above code uses modulo arithmetic and precomputes `power = base^n mod mod`, but actually for the outgoing character we need to subtract `charOut * (base^(n-1) mod mod) * base`? Wait, no. Actually, the common way is to compute the hash without storing the entire power value for base^n. Alternatively, we can precompute `base^(n-1) mod mod` and use it. The typical Rabin-Karp implementation uses:
```
hash = (hash - charOut * base^(n-1)) * base + charIn
```
So we should precompute `base^(n-1) mod mod`.

Corrected code:
```java
    int m = haystack.length();
    int n = needle.length();
    if (n == 0) return 0;
    if (m < n) return -1;
    long base = 26;
    long mod = (long)1e9+7;
    long patternHash = 0;
    long currentHash = 0;
    long power = 1;
    for (int i = 0; i < n; i++) {
        patternHash = (patternHash * base + (needle.charAt(i) - 'a' + 1)) % mod;
        currentHash = (currentHash * base + (haystack.charAt(i) - 'a' + 1)) % mod;
        if (i < n-1) {
            power = (power * base) % mod;
        }
    }
    if (currentHash == patternHash && haystack.substring(0, n).equals(needle)) {
        return 0;
    }
    for (int i = n; i < m; i++) {
        char out = haystack.charAt(i - n);
        char in = haystack.charAt(i);
        currentHash = (currentHash - (out - 'a' + 1) * power % mod + mod) % mod;
        currentHash = (currentHash * base + (in - 'a' + 1)) % mod;
        if (currentHash == patternHash && haystack.substring(i - n + 1, i + 1).equals(needle)) {
            return i - n + 1;
        }
    }
    return -1;
}
```

super30admin:
Strengths:
- You have thought about an advanced technique (rolling hash) to solve the problem efficiently.
- You included a substring check to avoid false positives due to hash collisions.
- You handled edge cases like empty needle and when haystack is shorter than needle.

Areas for Improvement:
- The rolling hash implementation has a fundamental error: the outgoing character should be multiplied by base^(n-1), not base^n. To fix this, you should precompute base^(n-1) and use that to subtract the outgoing character's contribution. Alternatively, you can use a standard double hashing technique with two moduli to reduce collisions.
- Using BigInteger is not efficient for this problem because the operations are slow and the space is not constant. Instead, use integer arithmetic with a large prime modulus to avoid overflow and reduce the hash to a manageable size.
- Consider using a simpler approach first, like the naive method (which is acceptable for small strings) or the KMP algorithm (which has O(m+n) time and constant space). For the problem constraints (strings up to 10^4), the naive solution might be acceptable in practice because worst-case scenarios are rare.
- The code structure is good, but variable names could be improved. For example, use `patternHash` instead of `pHash`, and `windowHash` instead of `cur`.
- Test your solution with various cases, including those where the hash might collide, to ensure correctness.

PR #907, opened 2026-03-17
solved

jeetusingh247:


super30admin:
### strStr() (Rabin_Karp_StrStr.java)
Your solution is well-implemented and demonstrates a good understanding of the Rabin-Karp algorithm. Here are some strengths and areas for improvement:

Strengths:
- You correctly handle edge cases (empty needle, haystack shorter than needle).
- The use of a rolling hash with base 26 and modulus 1000000007 is appropriate for lowercase English letters.
- You avoid hash collisions by verifying with a direct string comparison when hashes match.
- The code is clean, commented, and easy to follow.

Areas for Improvement:
- While the Rabin-Karp algorithm is efficient on average, it has a worst-case time complexity of O(m*n) due to collisions. For this problem, since the constraints are only lowercase English letters, it is acceptable. However, in an interview setting, you might be asked about worst-case scenarios. You could consider mentioning alternative algorithms like KMP for worst-case linear time, but your current solution is correct.
- The calculation of `power` is correct, but note that if n=0, your code returns 0 early, so no issue. However, if n=1, the loop for power runs 0 times, which is correct since base^(0)=1.
- There is a small potential issue: when computing the hash, you use `(haystack.charAt(i) - 'a' + 1)`. This maps 'a' to 1, 'b' to 2, etc. This is good to avoid zero values, but ensure that the modulus is large enough to minimize collisions. With mod=10^9+7, it is sufficient.
- In the window update step, you subtract the left character's contribution: `(haystack.charAt(i) - 'a' + 1) * power % mod`. This is correct, but note that `power` is base^(n-1) mod mod, which is exactly what is needed.

Overall, your solution is correct and efficient for the problem. Well done!

VERDICT: PASS

---

### String Anagrams  (Find_all_Anagrams.java)
Your approach to use a sliding window with a hashmap and a match counter is a good idea and can lead to an efficient O(m) solution. However, there are several issues in your implementation:

1. **Window Size**: You need to maintain a window of exactly `n` (the length of `p`). Your current code does not do this. You should only start checking for anagrams once the window has reached size `n`. Also, you should remove the character that is leaving the window when you move the window forward.

2. **Match Counter Logic**: The match counter should increment when a character's frequency in the window exactly matches its frequency in `p` (i.e., becomes zero), and decrement when it no longer matches (becomes non-zero). Your code does not correctly update the match counter when characters are added or removed.

3. **Indexing for Removal**: When `i >= n`, you should remove the character at `i - n` (the one leaving the window). But you are not correctly handling the case when the window hasn't reached size `n` yet.

4. **Checking Condition**: You should only check if `match == map.size()` when the window has exactly `n` characters. In your code, you check at every index `i`, which is incorrect for windows smaller than `n`.

Here is a corrected version of your approach:

```java
class Solution {
    public List<Integer> findAnagrams(String s, String p) {
        List<Integer> result = new ArrayList<>();
        int m = s.length();
        int n = p.length();
        if (m < n) return result;

        Map<Character, Integer> map = new HashMap<>();
        for (char c : p.toCharArray()) {
            map.put(c, map.getOrDefault(c, 0) + 1);
        }

        int match = 0;
        int left = 0;
        for (int right = 0; right < m; right++) {
            char in = s.charAt(right);
            if (map.containsKey(in)) {
                int count = map.get(in);
                count--;
                map.put(in, count);
                if (count == 0) match++;
            }

            // When the window size becomes greater than n, remove the leftmost character
            if (right - left + 1 > n) {
                char out = s.charAt(left);
                if (map.containsKey(out)) {
                    int count = map.get(out);
                    if (count == 0) match--;
                    count++;
                    map.put(out, count);
                }
                left++;
            }

            // Check if the current window is an anagram
            if (match == map.size()) {
                result.add(left);
            }
        }
        return result;
    }
}
```

However, note that the above code still has an issue: the window size is `right - left + 1`, and we only check when it is exactly `n`? Actually, we should check only when the window size is exactly `n`. So we can adjust:

```java
class Solution {
    public List<Integer> findAnagrams(String s, String p) {
        List<Integer> result = new ArrayList<>();
        int m = s.length();
        int n = p.length();
        if (m < n) return result;

        Map<Character, Integer> map = new HashMap<>();
        for (char c : p.toCharArray()) {
            map.put(c, map.getOrDefault(c, 0) + 1);
        }

        int match = 0;
        int left = 0;
        for (int right = 0; right < m; right++) {
            char in = s.charAt(right);
            if (map.containsKey(in)) {
                int count = map.get(in);
                count--;
                map.put(in, count);
                if (count == 0) match++;
            }

            // If the window size is larger than n, remove leftmost character
            if (right - left + 1 > n) {
                char out = s.charAt(left);
                if (map.containsKey(out)) {
                    int count = map.get(out);
                    if (count == 0) match--;
                    count++;
                    map.put(out, count);
                }
                left++;
            }

            // Check only when the window size is exactly n
            if (right - left + 1 == n && match == map.size()) {
                result.add(left);
            }
        }
        return result;
    }
}
```

This should work correctly. Keep in mind that the `map.size()` here is the number of distinct characters in `p`, and `match` counts how many of these distinct characters have their frequency exactly matched (i.e., count becomes zero). When `match` equals `map.size()`, the window is an anagram.

VERDICT: NEEDS_IMPROVEMENT

PR #908, opened 2026-03-25
Strings-2

nagasai67:


super30admin:
### strStr() (FindAllAnagrams.py)
Your solution for finding all anagrams is well-implemented and efficient for that problem. However, you have solved the wrong problem. The problem you were given is strStr(), which requires you to find the first occurrence of a substring (needle) in a string (haystack). 

For the strStr() problem, you need to implement a function that takes two strings: haystack and needle, and returns the index of the first occurrence of needle in haystack. A common approach is to use a nested loop to check for the substring at each possible starting index in haystack, which has a time complexity of O(m*n). Alternatively, you could use more advanced algorithms like Knuth-Morris-Pratt (KMP) for O(m+n) time, but the reference solution uses the simpler approach.

Please ensure you read the problem statement carefully and implement the correct function. Your coding skills are good, as seen in the anagram solution, but attention to the problem requirements is crucial.

VERDICT: NEEDS_IMPROVEMENT

---

### String Anagrams  (FirstOccurence.py)
Your implementation of the Rabin-Karp algorithm for substring search is well-done and efficient. However, there seems to be a misunderstanding: the problem you were asked to solve is "String Anagrams", which requires finding all starting indices of anagrams of string `p` in string `s`. An anagram is a permutation of the characters, not an exact substring. Your current solution only checks for exact matches.

To solve the correct problem, you should consider using a sliding window approach with frequency counts. Here's a suggested approach:
1. Create a frequency array for characters in `p`.
2. Use a sliding window of length `len(p)` over `s`.
3. Maintain a frequency array for the current window in `s`.
4. Compare the frequency arrays to check if the window is an anagram of `p`.
5. Adjust the window by moving one character at a time, updating the frequency array accordingly.

This approach would have a time complexity of O(m) and space complexity of O(1) (since the frequency arrays are of fixed size 26). 

You can also optimize by using a single frequency array and a counter to track the number of characters that match the frequency in `p`. This avoids comparing the entire frequency array each time.

Please revise your solution to address the correct problem. Your current code is good for substring search, but not for anagrams.

VERDICT: NEEDS_IMPROVEMENT

PR #909, opened 2026-03-26
strings - 2

shinjaneegupta:


super30admin:
### strStr() (Anagrams.py)
It appears there has been a misunderstanding. The problem you were asked to solve is strStr(), which involves finding the first occurrence of a substring (needle) in a string (haystack). However, your solution is for a different problem: "Find All Anagrams in a String". 

For the strStr() problem, you need to implement a function that returns the index of the first occurrence of needle in haystack, or -1 if not found. A common approach is to use a nested loop to check each possible starting position in haystack for the substring needle. Alternatively, you could use more advanced algorithms like KMP for better worst-case performance, but the naive approach is acceptable for many cases.

Here is a simple approach for strStr():
1. Get the lengths of haystack (m) and needle (n).
2. Iterate through haystack from index 0 to m - n (inclusive).
3. For each index i, check if the substring starting at i matches needle.
4. If it matches, return i.
5. If no match is found, return -1.

Example code in Python:
```python
class Solution:
        m = len(haystack)
        n = len(needle)
        for i in range(0, m - n + 1):
            if haystack[i:i+n] == needle:
                return i
        return -1
```

Please review the problem statement again and ensure your solution matches the requirements. Your current solution for the anagram problem is good, but it is not what was asked.

VERDICT: NEEDS_IMPROVEMENT

---

### String Anagrams  (FirstOccurrence.py)
Your solution demonstrates a good understanding of the rolling hash technique for substring search (strStr). However, there are a few points to note:
1. You are solving the "First Occurrence" (strStr) problem, but the problem assigned was "String Anagrams". Please read the problem statement carefully to ensure you are solving the correct problem.
2. For the strStr problem, your approach is valid, but to improve efficiency, you should precompute `base_n` (26^n) once instead of recalculating it every time you remove a character. This would make the inner loop constant time.
3. Also, consider using a modulus to avoid large numbers (e.g., a large prime) to prevent potential performance issues with big integers, though in Python it might not be critical.
4. For the anagram problem, a common efficient solution uses a sliding window with a frequency array. You can maintain a count of each character in the current window and compare it with the frequency of string p. This approach runs in O(m) time and O(1) space (since the array size is fixed at 26).

For example, for "String Anagrams", you can do:
- Create a frequency array for p.
- Use two pointers (left and right) to traverse s, and update a frequency array for the current window.
- When the window size equals len(p), check if the frequency arrays match (or use a counter variable to track matches).
- Then move the left pointer to slide the window.

Here's a sketch in Java (but you can adapt to Python):
```
public List<Integer> findAnagrams(String s, String p) {
    List<Integer> list = new ArrayList<>();
    if (s == null || s.length() == 0 || p == null || p.length() == 0) return list;
    int[] hash = new int[26];
    for (char c : p.toCharArray()) {
        hash[c - 'a']++;
    }
    int left = 0, right = 0, count = p.length();
    while (right < s.length()) {
        if (hash[s.charAt(right) - 'a'] >= 1) {
            count--;
        }
        hash[s.charAt(right) - 'a']--;
        right++;
        if (count == 0) {
            list.add(left);
        }
        if (right - left == p.length()) {
            if (hash[s.charAt(left) - 'a'] >= 0) {
                count++;
            }
            hash[s.charAt(left) - 'a']++;
            left++;
        }
    }
    return list;
}
```

VERDICT: NEEDS_IMPROVEMENT

PR #910, opened 2026-03-26
Problem2 added

megharaykar:


super30admin:
### strStr() (Problem2.py)
You have implemented a solution for "Find All Anagrams in a String" (LeetCode 438) instead of "strStr()" (LeetCode 28). This is a critical mistake. Please ensure you are solving the correct problem.

For the problem "strStr()", you need to find the exact first occurrence of the substring `needle` in `haystack`. The approach you used (sliding window with frequency counting) is suitable for anagram problems but not for exact substring matching.

To solve "strStr()", you can use a simple nested loop to check each possible starting index in `haystack` for the substring `needle`. Alternatively, you can use more advanced algorithms like Knuth-Morris-Pratt (KMP) for better efficiency in worst-case scenarios.

Here is a simple Java solution for reference:
```java
class Solution {
        int m = haystack.length();
        int n = needle.length();
        for (int i = 0; i <= m - n; i++) {
            if (haystack.substring(i, i + n).equals(needle)) {
                return i;
            }
        }
        return -1;
    }
}
```
Note: This solution has O(m*n) time complexity, which is acceptable given the constraints. However, for large inputs, you might consider using KMP for O(m+n) time.

VERDICT: NEEDS_IMPROVEMENT

---

### String Anagrams 
Your solution shows a good understanding of the sliding window approach and the use of a frequency map to track anagrams. However, there are a few issues to address:

- **Window Maintenance**: Your code attempts to remove the leftmost character when `i >= n`, but this removal should be done in a way that maintains a window of fixed size (equal to `len(p)`). Currently, you are processing the current character `s[i]` without having adjusted the window for the entire substring. This leads to incorrect frequency counts.

- **Match Condition**: The condition `match == len(hm)` is used to check if all characters in `p` are matched. However, `len(hm)` is the number of distinct characters in `p`, not the total count of characters. For example, if `p = "aab"`, then `len(hm)` is 2 (for 'a' and 'b'), but you need to have two 'a's and one 'b' matched. Your approach correctly tracks when each distinct character's frequency reaches zero, but you must ensure that the window size is exactly `n` when you check the match.

- **Initial Window**: Your code starts processing from the first character without having built a window of size `n`. This means that for the first `n-1` characters, you are incrementally building the frequency map but not checking for anagrams until the window is full. However, your code checks the match condition at every step from the beginning, which is incorrect.

- **Index Calculation**: When you find a match, you add `i-n+1` as the start index. This is correct only if you have maintained a window of size `n`. But due to the way you handle the window, this index might be off.

**Suggestions**:
1. Initialize the frequency map for `p` correctly (as you did).
2. Use two pointers (left and right) to represent the window, or iterate with right pointer and adjust left pointer when the window size exceeds `n`.
3. Alternatively, you can first build the frequency map for the first `n` characters of `s`, then slide the window by one step at a time, updating the frequency map and match count accordingly.
4. The match variable should be incremented when a character's frequency in the window matches exactly the frequency in `p` (i.e., becomes zero), and decremented when it no longer matches (becomes non-zero). But note: the frequency map for the window should be compared against the map for `p`.

Here is a corrected version of your approach:

```python
class Solution:
        res = []
        n = len(p)
        if len(s) < n:
            return res
        
        freq_p = {}
        for char in p:
            freq_p[char] = freq_p.get(char, 0) + 1
        
        freq_window = {}
        match = 0
        required = len(freq_p)
        
        left = 0
        for right in range(len(s)):
            char_right = s[right]
            if char_right in freq_p:
                freq_window[char_right] = freq_window.get(char_right, 0) + 1
                if freq_window[char_right] == freq_p[char_right]:
                    match += 1
            
            if right - left + 1 == n:
                if match == required:
                    res.append(left)
                
                char_left = s[left]
                if char_left in freq_p:
                    if freq_window[char_left] == freq_p[char_left]:
                        match -= 1
                    freq_window[char_left] -= 1
                left += 1
        
        return res
```

This version maintains a window of fixed size `n` and updates the match count correctly.

VERDICT: NEEDS_IMPROVEMENT

PR #911, opened 2026-03-27
Completed Strings-2

sarvanibaru:


super30admin:
### strStr() (FindAllAnagramsInString.java)
It appears you have submitted a solution for the wrong problem. The problem you solved is "Find All Anagrams in a String", but the problem you were assigned is "strStr()", which requires finding the first occurrence of a substring (needle) in a string (haystack). 

For the strStr problem, you need to implement a function that returns the index of the first occurrence of needle in haystack. You can use a simple nested loop approach (as in the reference solution) or more advanced algorithms like KMP for better worst-case performance. 

Please resubmit a solution for the correct problem. If you intended to submit for "Find All Anagrams in a String", then this solution is correct and efficient, but for the current context, it is not applicable.

VERDICT: NEEDS_IMPROVEMENT

---

### String Anagrams  (FindIndexOfFirstOccurenceInString.java)
Dear student,

It seems you have submitted a solution for the "Find Index of First Occurrence in String" problem (strStr) instead of the "String Anagrams" problem. Please double-check the problem you are solving and submit the correct code.

Even for the strStr problem, your Rabin-Karp implementation has a flaw in the rolling hash update. Specifically, when removing the outgoing character, you are subtracting `outCharVal = (outChar - 'a' + 1) * base^n`, but the correct exponent should be base^(n-1). For example, for a window of length n, the first character (the one leaving) was multiplied by 26^(n-1) in the current hash. So you should compute `posFac = base.pow(n-1)` and then subtract `outCharVal = BigInteger.valueOf(outChar - 'a' + 1).multiply(posFac)`. 

Also, note that using BigInteger might be inefficient for this problem due to the overhead. Instead, you can use a modulus to avoid overflow (e.g., a large prime number) to keep the hash within integer or long range.

For the actual "String Anagrams" problem, you need to return all start indices of anagrams of string `p` in string `s`. A common efficient solution uses a sliding window with two frequency arrays (or maps) to compare the counts. The Rabin-Karp technique is not typically used for anagrams because anagrams require frequency matching (which is not captured by a simple hash without collisions). However, if you want to use hashing, you would need a hash function that is invariant under permutation (e.g., the product of primes assigned to each letter) but that might lead to overflow.

I recommend implementing a standard sliding window approach for the anagram problem:
1. Create two frequency arrays of size 26 (for letters) for string `p` and the current window in `s`.
2. Initialize the frequency for `p` and the first window of length `p.length()` in `s`.
3. Compare the two frequency arrays. If they match, add the start index (0) to the result.
4. Then slide the window: decrement the count for the outgoing character (at start) and increment for the incoming character (at end). Compare the frequency arrays each time.

This approach has time complexity O(m) and space O(1).

VERDICT: NEEDS_IMPROVEMENT

PR #912, opened 2026-03-29
working solution

avcode3:


super30admin:
### strStr() (problem1.py)
Your solution is on the right track, but there is a critical error in the loop condition. You should change `while i < m - n` to `while i <= m - n` to ensure that you check all possible starting positions for the needle in the haystack. Additionally, you should add a check in the inner while loop to prevent going beyond the length of the haystack. For example, you could modify the inner while condition to `while k < m and j < n and haystack[k] == needle[j]` to avoid index errors.

Here is a corrected version of your code:
```python
class Solution:
        m = len(haystack)
        n = len(needle)
        if n == 0:
            return 0  # If needle is empty, return 0 as per standard?
        i = 0 
        while i <= m - n:  # Change to <=
            j = 0 
            if haystack[i] == needle[j]:
                k = i 
                while k < m and j < n and haystack[k] == needle[j]:
                    k += 1
                    j += 1
                    if j == n:
                        return i
            i += 1
        return -1
```

Note: The problem states that needle.length is at least 1, so the check for n==0 might not be strictly necessary, but it's good practice to handle edge cases.

Also, consider using more descriptive variable names for better readability. For example, `start_index` instead of `i`, `needle_index` instead of `j`, etc.

VERDICT: NEEDS_IMPROVEMENT

---

### String Anagrams  (problem2.py)
Your solution is efficient and correct, leveraging the sliding window technique to achieve optimal time complexity. Here are some suggestions for improvement:

1. **Variable Names**: Use more descriptive names. For example, `map_val` could be renamed to `freq_map` or `char_freq` to clearly indicate its purpose. `final_arr` could be `result` or `indices`.

2. **Edge Cases**: Consider explicitly handling the case where `n > m` by returning an empty list early. Although your code handles it implicitly (the loop won't run if `n > m` because `i` starts at 0 and `n` is larger than `m`, but actually `i` goes up to `m-1`, and the window size is `n`, so if `n > m`, the window never reaches size `n` and no indices are added), it's good practice to check for such cases for clarity.

3. **Initialization**: You initialize `match = 0`, which is correct. However, note that `len(map_val)` is the number of distinct characters in `p`, not the total count. This is appropriate because we need to match each distinct character's frequency to zero.

4. **Code Comments**: Adding comments to explain the sliding window process, especially the part where you remove the character that is leaving the window (when `i >= n`), would make the code more understandable.

5. **Efficiency**: Your solution is already optimal. However, you can avoid building the frequency map for `p` in a loop by using `collections.Counter` for clarity, but this is optional since the current method is efficient.

6. **Potential Bug**: There is a subtle issue: if `p` is empty, `len(map_val)` would be 0, and `match` would be 0, so the condition `match == len(map_val)` would be true for every index? But the problem states that `p.length >= 1`, so it's safe.

Here's a slightly revised version for clarity:
```python
class Solution:
        m, n = len(s), len(p)
        if m < n:
            return []
        
        freq = {}
        for char in p:
            freq[char] = freq.get(char, 0) + 1
        
        match = 0
        result = []
        
        for i in range(m):
            # Add current character
            if s[i] in freq:
                freq[s[i]] -= 1
                if freq[s[i]] == 0:
                    match += 1
            
            # Remove character that is out of the window
            if i >= n:
                out_char = s[i - n]
                if out_char in freq:
                    freq[out_char] += 1
                    if freq[out_char] == 1:
                        match -= 1
            
            # Check if the current window is an anagram
            if match == len(freq):
                result.append(i - n + 1)
        
        return result
```

Overall, your solution is excellent and outperforms the reference solution in terms of time complexity.

VERDICT: PASS

PR #913, opened 2026-04-06
Complete Strings-2

PrakarshKamal:


super30admin:
### strStr() (FindAllAnagrams.java)
It appears you have submitted a solution for the "FindAllAnagrams" problem instead of the "strStr" problem. Please check the problem statement and ensure you are solving the correct problem. For "strStr", you need to return the index of the first occurrence of the substring `needle` in the string `haystack`.

For the "strStr" problem, you can consider a simple approach by iterating through `haystack` and for each position, check if the substring starting there matches `needle`. This has a worst-case time complexity of O(m*n), which is acceptable given the constraints. Alternatively, you could implement more efficient algorithms like Knuth-Morris-Pratt (KMP) for O(m+n) time, but that is more advanced.

If you intended to solve "FindAllAnagrams", your solution is good: it efficiently uses a sliding window and a frequency map to track anagrams. However, for "strStr", you need to start over with the correct problem in mind.

VERDICT: NEEDS_IMPROVEMENT

---

### String Anagrams  (FirstIdxOccurence.java)
It seems you have submitted a solution for the "First Occurrence of a Substring" problem (strStr) instead of the "String Anagrams" problem. The problem you need to solve is to find all start indices in `s` where an anagram of `p` starts. 

For the "String Anagrams" problem, you need to:
1. Check if any substring of `s` of length equal to `p` is an anagram of `p`.
2. Return a list of all such start indices.

Your current solution implements three different approaches for substring search (naive, Rabin-Karp, and KMP), but these are for exact substring matching, not for anagram matching. 

To solve the anagram problem, you can consider:
- Using a frequency array or hashmap to count characters in `p` and then using a sliding window in `s` to check for anagrams.
- The sliding window approach can be optimized to O(n) time by maintaining a frequency count for the current window and comparing it with the frequency of `p`.

For example:
1. Create an array `pCount` of 26 integers to store the frequency of each character in `p`.
2. Create a similar array `sCount` for the current window in `s`.
3. Slide the window over `s`, and for each window, compare `sCount` with `pCount`. If they match, the window is an anagram.

This approach would be efficient with O(m + n) time and O(1) space.

Please ensure you are solving the correct problem. If you have code for the anagram problem, please submit that.

VERDICT: NEEDS_IMPROVEMENT

PR #914, opened 2026-04-10
 Done Strings-2

sainathek1999:


super30admin:
### strStr() (problem1.java)
Your solution is well-implemented and correctly addresses the problem. Here are some points to consider for improvement:
1. You have included a good edge case check for when the needle is empty (n==0), which is correct according to the problem constraints (though note that constraints say needle.length>=1, so this might be redundant but safe).
2. The while loop condition correctly checks boundaries for k and j, preventing index out-of-bounds errors.
3. The code is readable and follows standard naming conventions.

One minor optimization: Instead of using an inner while loop that increments k and j, you could use a for-loop to compare characters without modifying k and j directly. For example:
    for (int j = 0; j < n; j++) {
        if (haystack.charAt(i + j) != needle.charAt(j)) {
            break;
        }
        if (j == n-1) return i;
    }
This might be slightly more straightforward and avoids the need for extra variables (k). However, your current approach is also correct and efficient.

Overall, this is a solid implementation.

VERDICT: PASS

---

### String Anagrams  (problem2.java)
Your attempt to use a sliding window with a frequency map and a match counter is a good idea for efficiency. However, there are several issues:

- The frequency map should be used to track the difference between the current window and the pattern `p`. You start by storing the counts of characters in `p` as positive. Then, as you add a character to the window, you should decrement its count (indicating that you have "used" one occurrence). When the count becomes zero, it means you have exactly the required number of that character in the window. Similarly, when you remove a character, you increment the count. But your code does the opposite: it subtracts when adding (which is correct) but then when removing, it adds (which is also correct). However, the initial value is positive, and you subtract until it becomes zero or negative. The problem is that when you subtract and the count becomes zero, you increment `match`, which is correct. But when you remove a character, you add to the count. If the count was zero and becomes positive, you should decrement `match`. Your code does not do this; instead, it checks `if (cnt == 1) match--`, which is not symmetric. It should be: if the count was zero and becomes non-zero (specifically, becomes 1 when you are adding back), then decrement `match`. Similarly, when adding a character, if the count becomes zero from -1? Actually, the standard approach is to initialize the map with counts of `p`, and then for each character in the window, decrement. When the count becomes zero, increment `match`. When the count becomes non-zero (from zero) when removing, decrement `match`.

- You need to ensure that the window size is exactly the length of `p`. Your current code does not maintain a fixed window size. You should only check for an anagram when the window size is exactly `p.length()`. In your code, you are checking at every step after the window has been shrunk? Actually, you should check only when the window size is exactly `p.length()`. You can achieve this by having `right` from `0` to `s.length()-1`, and when `right - left + 1 == p.length()`, then check if `match == fmap.size()`.

- Consider using an array of size 26 instead of a hashmap since the problem states that the strings consist of lowercase English letters. This would be more efficient and simpler.

- Here is a corrected approach:
  - Initialize an array `count` of size 26 with the counts of `p`.
  - Initialize `left = 0`, `right = 0`, and `match = 0`.
  - For each character `c` in `s` at `right`:
        index = c - 'a'
        decrement `count[index]`
        if `count[index] == 0`, then increment `match`
        else if `count[index] == -1`, then decrement `match` (because it was zero and now becomes negative) — actually, no, you should only care about when it becomes zero and when it leaves zero. Actually, the standard way is to have `match` increment when a count becomes zero and decrement when a count becomes non-zero (from zero). So when you decrement and the count becomes zero, increment `match`. When you increment (when removing) and the count becomes non-zero (specifically 1), decrement `match`.

  - When the window size (`right - left + 1`) is greater than `p.length()`, remove the character at `left`:
        index = s.charAt(left) - 'a'
        increment `count[index]`
        if `count[index] == 1`, then decrement `match` (because it was zero and now becomes positive)
        else if `count[index] == 0`, then increment `match` (but this should not happen because you are incrementing from negative to zero? Actually, when you remove, you are adding back one occurrence. So if the count was -1 (meaning one extra), then after incrementing it becomes zero — which is good, so you should increment `match`? This is confusing.

  Alternatively, you can avoid using `match` and simply check if all counts are zero. But that would be O(26) per window, which is acceptable since 26 is constant.

  Given the complexity, I recommend using a fixed window size and checking the entire count array when the window size is reached. This would be O(26 * n) = O(n), which is efficient.

  Alternatively, you can use the match counter correctly. Here is a common solution:

        int[] count = new int[26];
        for (char c : p.toCharArray()) count[c-'a']++;

        int left = 0, right = 0, match = 0;
        List<Integer> res = new ArrayList<>();
        while (right < s.length()) {
            char c = s.charAt(right);
            count[c-'a']--;
            if (count[c-'a'] == 0) match++;
            else if (count[c-'a'] == -1) match--; // not necessary? Actually, we only care when it becomes zero and when it becomes non-zero from zero.

            if (right - left + 1 > p.length()) {
                char d = s.charAt(left);
                count[d-'a']++;
                if (count[d-'a'] == 0) match++;
                else if (count[d-'a'] == 1) match--;
                left++;
            }

            if (match == 26) { // wait, no: match should be the number of distinct letters that have count zero. But there are 26 letters, but not all are in p. Actually, we should only consider the distinct letters that are in p. But since we initialized with p, the counts for letters not in p are zero or negative? This is messy.

        }

  This is why many solutions use a simple check of the count array when the window is of size p.length(). They don't use a match counter.

  Given the above, I suggest for clarity to use the fixed window approach without a match counter, but with an array comparison. However, since the student aimed for O(n), they can use the match counter correctly by:

        int

VERDICT: NEEDS_IMPROVEMENT

PR #915, opened 2026-04-14
Strings-2

paridhimalviya:
Implement strStr() (https://leetcode.com/problems/implement-strstr/)
Find All Anagrams in a String (https://leetcode.com/problems/find-all-anagrams-in-a-string/)


super30admin:
### strStr() (FindAllAnagram.swift)
Your solution for the "Find All Anagrams" problem is well-implemented. You used a sliding window technique with a frequency map and match count, which is efficient and optimal for that problem. The code is clean and well-commented, which is good practice.

However, the problem you were asked to solve is "strStr()", which requires finding the first occurrence of a substring in a string. Your current code does not address this problem at all. It seems there might have been a misunderstanding or a mix-up with the problem.

For the "strStr()" problem, you need to implement a function that takes two strings: `haystack` and `needle`, and returns the index of the first occurrence of `needle` in `haystack`. You can use a simple nested loop to compare each possible starting index in `haystack` with `needle`, or more advanced algorithms like KMP for better worst-case performance (though the nested loop is acceptable for the constraints).

Please ensure you are solving the correct problem. For "strStr()", you should start by checking if `needle` is empty (return 0) and then iterate through `haystack` to see if `needle` matches starting at each index. Here's a simple approach in Swift:

```swift
func strStr(_ haystack: String, _ needle: String) -> Int {
    let n = haystack.count, m = needle.count
    if m == 0 { return 0 }
    if n < m { return -1 }
    
    for i in 0...(n - m) {
        let startIndex = haystack.index(haystack.startIndex, offsetBy: i)
        let endIndex = haystack.index(startIndex, offsetBy: m)
        let substring = haystack[startIndex..<endIndex]
        if substring == needle {
            return i
        }
    }
    return -1
}
```

This uses Swift's string comparison which might be efficient enough, but note that the worst-case time complexity is O(m*n). For a more efficient solution, you could implement the KMP algorithm.

VERDICT: NEEDS_IMPROVEMENT

---

### String Anagrams  (MatchThePattern_RK.swift)
Your code is well-structured and shows you understand substring search algorithms. However, you misunderstood the problem. The problem is about anagrams (permutations) of `p`, not exact matches. For example, if `p` is "abc", then "cba", "bac", etc. are valid anagrams. Your current solutions would not find these because they require the exact order "abc". 

To solve this problem, you should:
1. Use a frequency array (or dictionary) to count characters in `p`.
2. Use a sliding window of length `p.length()` in `s`, and maintain a frequency count for the window.
3. As you move the window, update the frequency count by removing the outgoing character and adding the incoming character.
4. Compare the frequency counts to determine if the window is an anagram of `p`.

This approach runs in O(m) time with O(1) space (since the frequency array size is fixed at 26 for lowercase letters).

VERDICT: NEEDS_IMPROVEMENT

PR #917, opened 2026-05-28
Strings 2

MeghaN28:


super30admin:
### strStr() (FindAllAnagrams.java)
**Major Issue**: You have submitted a solution for the wrong problem entirely. Your code solves "Find All Anagrams in a String" (LeetCode 438), but the assignment was to solve "strStr()" (LeetCode 28) - which asks to find the index of the first occurrence of needle in haystack, not find all anagram positions.

**Positive aspects of your code**:
- Excellent sliding window implementation
- Clear and comprehensive comments
- Good edge case handling
- Proper use of HashMap with getOrDefault()
- Clean variable naming and structure

**What you need to do**:
You need to implement the actual strStr() problem. The reference solution shows a straightforward approach: iterate through haystack, and for each position, check if the substring starting at that position matches the needle. Return the index when found, or -1 if no match is found after checking all possible positions.

VERDICT: NEEDS_IMPROVEMENT

---

### String Anagrams  (NeedleHaystackPrimecalc.java)
**
- **Strengths**: Clean code structure, good comments, efficient rolling hash implementation, proper handling of edge cases (empty needle, needle longer than haystack).
- **Weaknesses**: Solves the wrong problem entirely. The method signature and return type don't match the problem requirements.
- **Recommendation**: To fix this solution, you would need to:
  1. Change return type to `List<Integer>`
  2. Instead of comparing hash equality, check if current window is an anagram of p
  3. Collect all valid indices instead of returning on first match
  4. Use a sliding window with character frequency comparison rather than hash comparison

**

VERDICT: NEEDS_IMPROVEMENT

PR #918, opened 2026-05-28
find anagram,first index occurance of string

sadabala30:


super30admin:
## File 1: Index of the First Occurrence in a String.py

**Strengths:**
- Good implementation of the rolling hash (Rabin-Karp) concept
- Proper handling of negative hash values with modulo adjustment
- Collision detection by comparing actual strings

**Critical Issues:**
1. **Bug in power calculation**: `current_highest_power` is computed BEFORE the needle hash loop, so it equals 26^n instead of 26^(n-1). This causes incorrect hash removal during the rolling window. The power should be calculated as 26^(n-1) to properly remove the leading character's contribution.

2. **Edge case - empty needle**: If `needle` is empty, `prev_n` remains 0. The algorithm would return 0 (index of empty string), which may not be the intended behavior.

3. **Character encoding limitation**: Code assumes only lowercase letters (`ord(c) - ord('a') + 1`). Will fail for uppercase, digits, or special characters.

4. **Poor variable naming**: `prev_h`, `prev_n`, `current_highest_power` are cryptic and make the code hard to follow.

---

## File 2: findAnagrams.py

**Strengths:**
- Clean sliding window implementation
- Proper cleanup of zero-count entries from Counter
- Correct index calculation for result positions

**Issues:**
1. **Missing early return**: The code checks `if p_len > s_len` at the start, which is good, but the initial check happens after creating the first window unnecessarily.

2. **Inefficient Counter comparison**: Using `Counter` objects for every comparison is O(k) where k is alphabet size. For better performance, use integer arrays and track a `match_count` variable.

3. **Variable naming**: `p_len`, `s_len` are acceptable but could be more descriptive.

---

**Time Complexity:**
- strStr: O(m × n) worst case due to collision verification, O(m + n) average
- findAnagrams: O(s) where s = len(s)

**Space Complexity:**
- strStr: O(1) - constant extra space
- findAnagrams: O(1) - fixed 26-character alphabet (or O(k) for unique characters)

PR #919, opened 2026-05-29
adding string-2 solutions

nikhylw:


super30admin:
### strStr() (find_all_anagram.py)
**Strengths:**
- The code is well-structured and follows good coding practices
- The sliding window implementation is correct for its intended problem
- Variable naming is clear and descriptive
- Time and space complexity comments are included
- Proper handling of frequency map updates

**Critical Issue:**
- This solution solves the wrong problem entirely. The code implements "Find All Anagrams in a String" (LeetCode 438), not "strStr()" (LeetCode 28). The strStr problem requires finding the first occurrence of a substring, returning a single index or -1, not a list of all anagram positions.

**What needs to be fixed:**
1. Implement the correct algorithm for strStr() - either brute force (as in reference) or KMP for optimization
2. Return a single integer index, not a list
3. Handle the case where needle is empty (should return 0 per some definitions)

VERDICT: NEEDS_IMPROVEMENT

---

### String Anagrams  (strstr.py)
- **Strengths**: 
  - Good implementation of rolling hash technique
  - Clean, readable code structure
  - Correct space complexity
  - Proper handling of edge cases (empty needle, etc.)

- **Areas for Improvement**:
  - **Critical Issue**: The solution solves the wrong problem. It finds exact substring matches (strStr) instead of anagram matches (findAnagrams).
  - The problem requires checking character frequency/counts, not hash values of entire substrings.
  - For anagram matching, you need to compare character frequency distributions, not hash equality.
  - Consider using a sliding window with character count arrays to properly solve this problem.

VERDICT: NEEDS_IMPROVEMENT

PR #920, opened 2026-05-30
v

spencerkrebs:


super30admin:
### strStr() (anagrams-in-string.py)
**Major Issue:** You have submitted a solution for the wrong problem. The strStr() problem requires finding the index of the first occurrence of `needle` in `haystack`, not finding anagram positions. Your solution finds all starting indices where `p` is an anagram of a substring in `s`.

**What you need to do:**
- Implement a solution that searches for the exact `needle` string within `haystack`
- Return the index of the first occurrence, or -1 if not found
- Consider using built-in methods like `indexOf()` or implementing a simple substring matching algorithm

**Strengths of your code (for the anagram problem):**
- Clean sliding window approach
- Good use of hash map for frequency tracking
- Efficient O(n+m) time complexity
- Well-structured with clear variable names

**For the strStr problem, consider:**
- A simple brute force approach checking each position
- The KMP algorithm for O(m+n) time complexity
- Using Java's `indexOf()` method for simplicity

VERDICT: NEEDS_IMPROVEMENT

---

### String Anagrams  (index-first-occurence.py)
The student appears to have confused this problem with the classic "Implement strStr" (substring search) problem. To solve the String Anagrams problem correctly, the approach needs to be fundamentally different:

1. **Character Frequency Approach**: Use a frequency count array (size 26) to track character occurrences
2. **Sliding Window**: Instead of checking each position independently, slide a window of size n (length of p) across s
3. **Comparison**: Compare frequency counts to determine if current window is an anagram

The current solution would need complete restructuring. The student should:
- Use an int[26] array to count characters in p
- Use a sliding window to count characters in each substring of s
- Compare counts to check for anagrams
- Add all valid indices to the result list

VERDICT: NEEDS_IMPROVEMENT

PR #922, opened 2026-06-02
Done Strings-2

samikshm:


super30admin:
**find_all_anagrams_in_string.java:**
- **Correctness**: The sliding window approach with frequency map is correct for finding anagrams. The logic for tracking matched characters via the `match` counter is sound.
- **Time Complexity**: O(n + m) is stated correctly. The solution efficiently processes each character once.
- **Space Complexity**: O(k) where k is the number of unique characters in p is correct.
- **Code Quality**: Good use of HashMap and clear variable names. However, the code could benefit from extracting the character frequency building logic into a helper method for reusability. The comments are helpful but could be more concise.
- **Edge Cases**: Handles the edge case of s.length() < p.length() appropriately. Consider adding early return when p is empty (though this may not be a valid test case).

**strStr.java:**
- **Correctness**: The Rabin-Karp rolling hash implementation is conceptually correct but has a critical flaw - it lacks modulo arithmetic. Without modulo, hash values will overflow for long strings since `26^n` grows extremely fast. This will cause incorrect behavior or runtime errors for longer inputs.
- **Time Complexity**: O(m + n) is stated correctly for the average case, but worst case is O(m*n) due to hash collisions.
- **Space Complexity**: O(1) is correct.
- **Code Quality**: The rolling hash update logic is clear. However, the implementation is incomplete without modulo handling. Using `Math.pow()` for positional factor is inefficient - should use precomputed multiplication.
- **Edge Cases**: Doesn't handle empty needle string (would cause division by zero or incorrect behavior). Also doesn't verify actual string match after hash match to handle collisions.

**General Observations:**
- Both solutions include helpful comments explaining the approach.
- Variable naming is descriptive (in/out for sliding window characters).
- Code structure is consistent with standard LeetCode format.

PR #923, opened 2026-06-19
Complete Strings-2

dhruvil15:


super30admin:
### strStr() (Problem97.java)
**Strengths:**
- Good initiative to use an optimized approach (rolling hash)
- Clean, readable code structure
- Correct space complexity analysis
- Good variable naming conventions

**Areas for Improvement:**
1. **Critical Bug**: Without modulo arithmetic, hash values will overflow for longer strings. You must use modulo (e.g., `% 1000000007`) to keep values in bounds.

2. **Critical Bug**: Hash collisions are not handled. After a hash match, you MUST verify the actual substring character-by-character to ensure correctness.

3. **Algorithm Correction**: Either:
   - Add modulo arithmetic and string verification, OR
   - Revert to the simpler sliding window approach like the reference solution

The rolling hash technique is a valid optimization but requires careful implementation to avoid overflow and collision issues.

VERDICT: NEEDS_IMPROVEMENT

---

### String Anagrams  (Problem98.java)
**Strengths:**
- Good use of sliding window technique, which is more optimal than the reference solution
- Clean code structure and readable variable names
- Proper handling of the window sliding mechanism (adding new char, removing old char)

**Areas for Improvement:**
1. The `match` counter logic is incorrect. Consider using a simpler approach: track how many characters have zero frequency (all matched) instead of trying to increment/decrement match on negative frequencies.
2. The condition `match == map.size()` is flawed because `map.size()` only counts distinct characters in `p`, not whether all their frequencies are zero.
3. A cleaner approach would be: maintain a counter for "satisfied" characters (those with frequency 0), and check if satisfied == number of distinct characters in p.
4. Consider using an array of size 26 instead of HashMap for better performance, as we're dealing with lowercase English letters.

**Recommended Fix:**
```java
int matched = 0;
int distinct = map.size();

for (int i = 0; i < m; i++) {
    // add character at i
    if (map.containsKey(in)) {
        map.put(in, map.get(in) - 1);
        if (map.get(in) == 0) matched++;
    }
    
    // remove character at i-n
    if (i >= n) {
        if (map.containsKey(out)) {
            if (map.get(out) == 0) matched--;
            map.put(out, map.get(out) + 1);
        }
    }
    
    if (i >= n - 1 && matched == distinct) {
        result.add(i - n + 1);
    }
}
```

VERDICT: NEEDS_IMPROVEMENT

